PR #1506, opened 2015-01-27
Feature/doc fixes: Deprecation notice for link walking

kesslerm:
Confusion about the status of link walking in Riak-2.0 has surfaced several times over the last week. I have resurrected the deprecation warning that was dropped without an explanation.

Please review and merge if appropriate


kesslerm:
See https://github.com/basho/riak-java-client/issues/494 for a recent example. Brian Roach supplied the information about support in 2.0 there.

PR #1511, opened 2015-01-29
Document limitations of using the bitcask sync interval feature

danabr:


danabr:
I noticed now that the same thing would happen if setting bitcask_merge_check_interval to a value lower than 60 seconds. It is probably less likely that would happen, though.

laurenrother:
@danabr I'm super sorry that I didn't see this until just now. But it looks good! So :candy:

Issue #1547, opened 2015-03-05
Fast track layout and menu broken [JIRA: PMT-100]


roooms:
http://docs.basho.com/riak/latest/tutorials/fast-track/ 
* menu and page structure is broken
* the nav on the left seems to append urls together incorrectly, causing 404’s

kesslerm:
The latest working version of this belongs to Riak 1.3.2. Is there a left-over file in the 'latest' section on the doc-server?

This one works as intended: http://docs.basho.com/riak/1.3.2/tutorials/fast-track/

This one shows a 404 error, as intended: http://docs.basho.com/riak/2.0.5/tutorials/fast-track/

roooms:
The link to this URL is available in the FAQs at http://docs.basho.com/riak/latest/community/faqs/basics/#what-is-riak

TheOnlyRew:
With #1549 merged, can this issue be closed?

laurenrother:
Please feel free to reopen this issue if the PR above did not satisfactorily resolve the issue.

PR #1549, opened 2015-03-05
Do not link to outdated fast track tutorial

kesslerm:
The fast track tutorial only existed until Riak-1.3. Use
more recent documents instead. 

Further small edits to the faqs/basics.html.fml file

Addresses issue #1547


kesslerm:
+1 from @nerophon via private communication

PR #1551, opened 2015-03-06
Be more explicit about using riak attach to verify the poodle patch

macintux:
Clarify both the trailing period and that for 1.x products we use `^C a` to exit an attached shell.


mjbrender:
Looks helpful. Merging unless you see anything further @macintux

PR #1552, opened 2015-03-10
Copying text regarding which configuration file to use for MDC to the MDC config docs

drewkerrigan:
Currently the v3 MDC config docs only mention app.config even though users actually need to use `advanced.config`.


dmitrizagidulin:
+1, looks good.

kesslerm:
To be consistent with source/languages/en/riak/ops/mdc/v3/aae.md, please also change the code blocks to have the advancedconfig annotation instead of appconfig.

laurenrother:
@kesslerm Does my update address everything? Or do I also need to change the other code blocks?

Issue #1553, opened 2015-03-11
Bitcask merge policy documentation is incorrect [JIRA: PMT-59]


nerophon:
http://docs.basho.com/riak/latest/ops/advanced/backends/bitcask/#Configuring-Bitcask

...and scrolling down to the section on Merge Policy. The third policy option is listed as:

"Start and end time — Hours during which merging is permitted"

However, in fact, you must specify the policy as "window", as well as including a start and end time, e.g.

```
bitcask.merge.policy=window
bitcask.merge.window.start=23
bitcask.merge.window.end=01
```

Please see also:
https://basho.zendesk.com/agent/tickets/10481

dmitrizagidulin:
Fixed in https://github.com/kesslerm/basho_docs/commit/2257469072dc2d3ec4efc7e391834e7bb2ec02e6 , merged and deployed live.
Closing.

dmitrizagidulin:
Ah, whoops, looked in the wrong place. Still have to merge.

PR #1554, opened 2015-03-11
Document the use of the window merge policy better

kesslerm:
The 'window' merge policy documentation was still half-way
between the previous app.config style documentation and the
new riak.conf style documentation. Use riak.conf style primarily
from now on.

Addresses https://github.com/basho/basho_docs/issues/1553


jhmullen:
+1

PR #1555, opened 2015-03-12
add signature_v2 on s3cmd's config

ksauzz:
s3cmd was changed to use AWS Authentication v4 on default, but current RIak CS release doesn't support v4 yet. So, `signature_v2 = True` is require to connect to Riak CS via s3cmd.


kuenishi:
+1

PR #1556, opened 2015-03-12
Fixed mispelling of 'dropped'

sargun:


mjbrender:
Obvious clean up - thanks @sargun!

PR #1558, opened 2015-03-14
Expand the Riak Metrics to Graph section

dmitrizagidulin:
Users frequently ask "So which metrics should I graph?". This is an attempt to expand the Metrics to Graph section and provide some guidance for choosing a "bare minimum" set of stats (from various Professional Services recommendations and training documents).

### DIFF
--- a/source/languages/en/riak/ops/running/stats-and-monitoring.md
+++ b/source/languages/en/riak/ops/running/stats-and-monitoring.md
@@ -13,28 +13,147 @@ moved: {
 ---
 
 Riak provides data related to current operating status, which includes
-statistics in the form of counters and histograms. These statistics are
-made available through the HTTP API via the `[[/stats|HTTP Status]]`
-endpoint, or through the `[[riak-admin|riak-admin Command Line]]`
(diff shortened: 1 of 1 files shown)


drewkerrigan:
+1 3873be9, Much more useful in this format as well. Eventually we should add links to the monitoring plugins after they are polished.

TheOnlyRew:
It looks like this commit is inconsistent about column lengths. I think Luc was pretty strict about a 70-column limit where possible, and -- while the tables shouldn't bother with the limit -- it looks like there's a good bit of prose that should be brought in.

dmitrizagidulin:
@TheOnlyRew thanks! I'll revise to fall under 70-col limit. Can we add that requirement to the README?

TheOnlyRew:
That's a really good idea, actually. I'll try to walk the docs a bit and see if there are any other style guidelines that should be written down.

dmitrizagidulin:
Revised to be under 80-col limit, as per discussion in Docs room.

TheOnlyRew:
+1

PR #1560, opened 2015-03-16
Update version compatibility on Riak CS and Riak

kuenishi:


kuenishi:
There is another page: http://docs.basho.com/riakcs/latest/cookbooks/Version-Compatibility/ Looks like a duplicate.

TheOnlyRew:
It looks like source/languages/en/riakcs/cookbooks/installing/Installing-Riak-CS.md and source/languages/en/riakcs/cookbooks/Version-Compatibility.md duplicate information, but present different contexts---the former is an install guide, the later a reference for easy searching/linking.

I'd say leave the duplication and update source/languages/en/riakcs/cookbooks/Version-Compatibility.md as part of this patch as well.

Issue #1561, opened 2015-03-16
[Documentaion] broken links in "The Riak API" docs [JIRA: PMT-76]


iStefo:
The paragraph on http://docs.basho.com/riak/1.1.0/references/appendices/concepts/#The-Riak-APIs contains broken links. I know that this is an issue of very low priority, but I think it should be said.

kesslerm:
Checked links on page; these seem to be broken:

http://blog.basho.com/2010/04/27/hello-bitcask/

Not resolved from Markdown:  
[[a REST API]] 
[[Protocol Buffers API]]

http://webmachine.basho.com/ could be a direct link to the github page instead

dmitrizagidulin:
Sounds like we need to change the following links:
1. http://blog.basho.com/2010/04/27/hello-bitcask/ to http://basho.com/hello-bitcask/
2. http://webmachine.basho.com/  to https://github.com/basho/webmachine
3. ```[[a REST API]]``` to ```[[a REST API|http://docs.basho.com/riak/1.1.0/references/apis/http/]]
4. ```[[Protocol Buffers API]]``` to ```[[Protocol Buffers API|http://docs.basho.com/riak/1.1.0/references]]```/apis/protocol-buffers/

TheOnlyRew:
This looks to be fixed. Please reopen if I'm wrong.

Basho-JIRA:
See above comment for resolution. 

***[posted via JIRA by Lauren Rother]***

PR #1562, opened 2015-03-16
Fix broken links to bitcask blog article

kesslerm:
Addresses comment in https://github.com/basho/basho_docs/issues/1561


TheOnlyRew:
Looks good to me.
Should we mark this PR as one that will need to be back-ported to previous releases?

PR #1565, opened 2015-03-17
Obvious spelling error is obvious...length shortened

tylerhannan:
The promo banner has always had length issues...but the spelling error was egregious.

This attempts to rectify both (without actually making the banner fully responsive because...reasons).


TheOnlyRew:
Thanks for catching this Tyler. I don't know if I ever would have noticed if it wasn't brought up, and now I can't un-see it...

Issue #1584, opened 2015-03-23
Setting suggestions duplicated in tuning docs [JIRA: PMT-103]


roooms:
Overlap and inconsistency of sysctl tcp settings specified twice on the same page at [Storage and File System Tuning](http://docs.basho.com/riak/1.4.12/ops/tuning/linux/#Storage-and-File-System-Tuning) and [Kernel and Network Tuning](http://docs.basho.com/riak/1.4.12/ops/tuning/linux/#Kernel-and-Network-Tuning).

laurenrother:
Please feel free to reopen this issue if the above PR did not address your concerns.

PR #1585, opened 2015-03-23
Fix network tuning settings being in the wrong section

kesslerm:
Addresses https://github.com/basho/basho_docs/issues/1584

Additional network tuning sysctl settings were originally introduced in
6f5f31242b9ba02e16395f554ee47cf45af8610c, but seem to have
ended up in the wrong section due to a merge gone wrong.


TheOnlyRew:
I thought the intention of that page was to have some duplication.
The change makes sense, though. +1

Issue #1194, opened 2014-07-16
rbenv can't handle 'default' keyword in .ruby-version


ksauzz:
Although there are some workarounds, I hope various ruby version managers would work fine here against the version in .ruby-version.

My suggestions are:
- option1: put specific ruby version into .ruby-version
- option2: remove .ruby-version

```
kazuhiro@chipmunk☁ /usr/local/src/basho_docs  [master]
%
rbenv: version `default' is not installed
kazuhiro@chipmunk☁ /usr/local/src/basho_docs  [master]
% rbenv version
rbenv: version `default' is not installed
 (set by /usr/local/src/basho_docs/.ruby-version)
rbenv: version `default' is not installed
```

rbenv can't handle and install 'default' as ruby version without using some trick.

PR #1589, opened 2015-03-25
Starter's grief

jpease:
Probably good to have a defined known working version of Ruby. Tried running against 2.2 and there are a number of gem issues that would need to be resolved.

Also would've addressed this issue:
https://github.com/basho/basho_docs/issues/1194


TheOnlyRew:
I'm not much of a Ruby developer, but this change looks good to me, especially as 2.2 presents issues.
+1

mjbrender:
Hey @bkerley, would defining the Ruby version in https://github.com/basho/basho_docs/commit/dbcbef69c55abe3c91e79f59299a2d802abf8c05 cause any unforeseen issues you're aware of?

bkerley:
It shouldn't break anything.

@jpease FWIW I was able to make it work against 2.2.1 with a `bundle update`.

PR #1592, opened 2015-03-25
Add riak-cs.conf as an autogenerated codeblock.

TheOnlyRew:
Codeblocks such as
\`\`\`riakcsconf
rewrite_module = riak_cs_s3_rewrite
\`\`\`
now render correctly with the title showing `riak-cs.conf`.


jhmullen:
+1

Issue #1577, opened 2015-03-18
Strange first few lines in Recovery From Backups [JIRA: PMT-90]


nerophon:
http://docs.basho.com/riak/latest/ops/running/recovery/failure-recovery/#Cluster-Recovery-From-Backups

The first few lines of this section are messed up.

PR #1600, opened 2015-03-30
fixing confusing lines from merge

jhmullen:
this addresses issue #1577


TheOnlyRew:
Lines 119 and 120 should also be removed, I think.

kesslerm:
The warning in lines 119:120 can be folded into the start of the section.

TheOnlyRew:
Step 3 is already "Ensure that Riak is not running on any of the nodes." so the warning is pretty redundant.

TheOnlyRew:
+1

PR #1601, opened 2015-03-31
Update strong-consistency.md

jrnt30:
Updating strongly consistent read example to reflect the intention that every read will consistently return the "v2" version of the Man-U.


laurenrother:
:candy:

Issue #1606, opened 2015-04-02
Reference changed legacy app.config settings in upgrade doc and release notes [JIRA: PMT-131]


jhmullen:
```
 %% Riak CS http/https port and IP address to listen at
%% for object storage activity
{listener, {"127.0.0.1", 8080}},
%%{cs_ip, "127.0.0.1"},
%%{cs_port, 8080 } ,

%% Riak node to which Riak CS accesses
{riak_host, {"127.0.0.1", 8087}},
%%{riak_ip, "127.0.0.1"},
%%{riak_pb_port, 8087 } ,

%% Configuration for access to request
%% serialization service
{stanchion_host, {"192.168.33.10", 8085}},
%%{stanchion_ip, "192.168.33.10"},
%%{stanchion_port, 8085 },
```

These settings have changed syntax and need to be mentioned in the docs.

jhmullen:
Merged and Deployed

PR #1607, opened 2015-04-02
Add warnings about changed settings in legacy app.config

jhmullen:
This addresses issue #1606 (PMT-131) in the Rolling Upgrades and config reference documentation, though the Release notes should be updated as well.


TheOnlyRew:
I've reviewed the two commits that @jhmullen contributed so far, and they look good to me.

PR #1610, opened 2015-04-04
Added missing brace to fix download link

alasdairnicol:
This should fix the broken link in the "Installing from source package" section of the docs.

http://docs.basho.com/riak/2.0.5/ops/building/installing/from-source/


laurenrother:
:candy:

PR #1611, opened 2015-04-05
Fix malformed tag in cassandra comparison

jergason:
Looks like the `<td>` was missing an opening angle bracket.


TheOnlyRew:
Thanks so much for the bug fix!
Unfortunately, when I applied you patch locally to take a look at the page, there were a few other issues with the html tags used, with the table width, and ohmygod the formatting in that file... I couldn't help myself but to fix up whitespace, close out a few mistaken tags, and covert the few table-breaking markdown links to anchors.
I don't mean to steal anyone's thunder, but I've rolled your change into #1624 along with the above. As such, I'm going to close out this PR un-merged.

Thanks for bringing this to our attention, though!

PR #1624, opened 2015-04-09
Fix tags, formatting, links in Cassandra comparison.

TheOnlyRew:
Based on @jergason discovery from #1611.


laurenrother:
:cake:

PR #1631, opened 2015-04-10
[ops/advanced/security] Remove explicit app.config from text

kesslerm:
Examples are given for both app.config and riak.conf, and
the text body should no longer spell out app.config as the
preferred way to configure the erlang distribution port range.


jhmullen:
+1

PR #1639, opened 2015-04-14
Add a description of minus in version range.

ksauzz:
This behaviour is defined [here](https://github.com/basho/basho_docs/blob/0e834b7ab8ae9ef1c0dd937b7ec2320d62f2afdb/lib/tools.rb#L20-L29).

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ version). The ranges are specified using either greater/less than or
 plus/minus signs or a version range.
 
 * **{{1.0.0+}}** _(greater than 1.0.0, inclusive)_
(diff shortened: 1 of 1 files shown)


TheOnlyRew:
+1
This can be merged in at any time but, as it's not part of the 2.1.0 push, we may wait until Thursday.

ksauzz:
Agree. Thanks :)

TheOnlyRew:
@laurenrother would mind taking a look at this and merging it in to master?

laurenrother:
:candy:

PR #1640, opened 2015-04-14
[upgrade to 2.0.x <= 2.0.5] Document need to set default bucket properti...

kesslerm:
...es

If app.config is used instead of riak.conf, the
'allow_mult' and 'dvv_enabled' properties are not set as
described in the rest of the documentation. As a workaround
one can set these explicitly in 'app_config'.

This is a workaround for issue https://github.com/basho/riak/issues/727


jhmullen:
+1

Issue #1634, opened 2015-04-13
Create Devrel Instructions for 2.1.0 [JIRA: PMT-148]


jhmullen:
In addition to the dmg instructions in Issue #1568 , @tylerhannan is working on devrel instructions for osx.

Issue #1568, opened 2015-03-17
Check/update links for chef/dmg installers [JIRA: PMT-80]


jhmullen:
Should target branch releases/riak_2.1.0

dmg installer content being generated by @drewkerrigan and @glickbot

jhmullen:
Changed title of issue, as Ansible was dropped for Q1

jhmullen:
Changed issue again as Puppet is also dropped for Q1

seemaj:
Add a link to https://github.com/basho-labs/puppet-riak here 
http://docs.basho.com/riak/latest/ops/building/installing/
for the Puppet installer

seemaj:
FYI Chef and Puppet are still at 2.0.5

seemaj:
Chef installer will be updated post release. However, we should remove the page for Chef http://docs.basho.com/riak/latest/ops/building/installing/with-chef/
And just link to the repo here http://docs.basho.com/riak/latest/ops/building/installing/
This needs to be updated for 2.0.5

For the dmg installer we agreed that we will create sections here http://docs.basho.com/riak/latest/quickstart/

1) dmg installer, description of what it is, link to https://github.com/basho-labs/riak-app 
2) devrel installer, description of what it is, link to https://github.com/basho-labs/riak-dev-cluster/
3) rest of instructions as normal to do it the old-fashioned way

seemaj:
We need to update docs when 
1. DMG installer is updated for 2.1 (Drew K)
2. Devrel installer is updated for 2.1 (Tyler)
3. Chef and Puppet are updated for 2.1 (Matt)

tylerhannan:
I will address separately, but updating the [Installing on OS X ](http://docs.basho.com/riak/latest/ops/building/installing/mac-osx/) in the Ops/Building section is probably necessary as well.

Re. Seema's list:

- [x] DMG Installer is updated for 2.1 (Drew K)
- [x] Devrel installer is updated for 2.1 (Tyler)
- [ ] Chef and Puppet are updated for 2.1 (Matt)

laurenrother:
Community installers will be linked in the community section of the docs as we get them. Closing this issue.

Basho-JIRA:
Community installers will get linked in the community section as they come in. 

***[posted via JIRA by Lauren Rother]***

PR #1643, opened 2015-04-14
Creating new subheads for dmg and devrel

jhmullen:
This branch and PR are meant to address issue #1568 and issue #1634 (PMT-148) .  It adds new subheads on the quickstart page to link to the dmg and devrel installers.

I have also added a Note, and gratuitous "OSX" warnings, so that new users can differentiate between the installers for testing vs the rest of the page for general purpose installation.


jhmullen:
Closing this because I targeted the wrong branch.

PR #1647, opened 2015-04-15
[upgrade-v20] clarify 'allow_mult' setting

kesslerm:
710a32be8b11502c404e3d08167875d0a9cafef1 added advice to
add additional settings for the default bucket type to
app.config. If the user had previously set 'allow_mult=true'
they should keep that setting.

Fixed the markup of the app.config fragment and a typo.


kesslerm:
Giving myself a +1, this got lost in the scramble to get the release out.

PR #1655, opened 2015-04-16
Add app.config upgrade issue to Product Advisories

laurenrother:


source/languages/en/riak/community/product-advisories/appconfig-forward.md
```diff
@@ -0,0 +1,29 @@
...
+
+Info | Value
+:----|:-----
+Date issued | April 15, 2015
+Product | Riak
+Affected Riak versions | 2.0.0+, 2.1.0+
+
+## Overview
+
+A forward incompatibility of app.config between 1.x and 2.x has been identified.
```
TheOnlyRew: Oddity of Basho; I've gotten the feeling that rather than organizing our version numbers `major.minor.patch`, the first two numbers are the major and the last is the minor (`ma.jor.minor`?). This line should probably read  
`A forward incompatibility of app.config between 1.4.x and 2.x has been identified.`

This note applies to the below paragraphs as well.

TheOnlyRew: Actually... This probably does express when upgrading from 1.3.x to 2.0.x...

I don't know if any of our customers have tried that, though. Hopefully none will, but my note might be incorrect in its recommendation.

source/languages/en/riak/community/product-advisories/appconfig-forward.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Description
+
+There is a forward compatibility issue when upgrading from the 1.x series to the 2.x series and continuing to use an app.config instead of migrating to the new riak.conf introduced in 2.0. This issue results in different default bucket properties being applied to un-typed buckets in 2.x, specifically for the `allow_mult` and `dvv_enabled` parameters. In 1.x `allow_mult` defaulted to 'false' but in 2.x instead defaults to 'true'. The `dvv_enabled` parameter did not exist in 1.x, and defaults to 'true' in 2.x.
```
TheOnlyRew: Should 'app.config' and 'riak.conf' be wrapped in backticks? I think I've seen some pages do that, but there are probably others that don't. 

Do we care enough to decide on a consistent style, and use that here + going forward?

TheOnlyRew: ```defaulted to `false` ``` would likely be more correct here, though.

source/languages/en/riak/community/product-advisories/appconfig-forward.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Impact
+
+A switch from an `allow_mult` of 'false' to an `allow_mult` of 'true' can cause unbounded growth of objects as siblings are added each time the object is updated unless the client application is explicitly resolving siblings.
+
+## Mitigation Strategy
+
+This issue can be mitigated by adding `{allow_mult, false}` to `default_bucket_props` in the `riak_core` section of the app.config. `{dvv_enabled, false}` can also be added, however enabling Dotted Version Vectors should have no negative impact on the cluster.
```
TheOnlyRew: We should probably write out a more complete example here, as the full configuration usually looks like
```
  {riak_core, [
    . . .
    {default_bucket_props, [
      {allow_mult, false},
      {dvv_enabled, false}]},
    . . .
```

source/languages/en/riak/community/product-advisories/appconfig-forward.md
TheOnlyRew: After thinking about it some, I think we should just add the second `.x`, so `1.x` would become `1.x.x`. It's kinda weird, but so is our versioning system.

source/languages/en/riak/community/product-advisories/appconfig-forward.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Description
+
+There is a forward compatibility issue when upgrading from the 1.x series to the
+2.x series and continuing to use an app.config instead of migrating to the new
+riak.conf introduced in 2.0. This issue results in different default bucket
+properties being applied to un-typed buckets in 2.x, specifically for the
+`allow_mult` and `dvv_enabled` parameters. In 1.x `allow_mult` defaulted to
+`false` but defaults to `true` in 2.x. The `dvv_enabled` parameter did not exist
+in 1.x, and defaults to 'true' in 2.x.
```
TheOnlyRew: in the line above, you have ```defaults to `true` ```. Are the `'`s here intentional?

TheOnlyRew:
+1

PR #1659, opened 2015-04-17
Change description of ETS acronym

derek121:
It had been referred to as "Erlang Table Storage". The following
resources describe it as "Erlang Term Storage":

http://docs.basho.com/riak/latest/ops/running/nodes/inspecting/
Descripion of "memory_ets", under "CPU and Memory"

http://docs.basho.com/riak/latest/ops/running/recovery/errors/
Descripion of "{system_limit, {ets, new}}", under "Lager Formats"

https://github.com/erlang/otp/blob/maint/lib/stdlib/src/ets.erl#L22
"Erlang Term Store"

https://www.safaribooksonline.com/library/view/programming-erlang-2nd/9781941222454/f_0155.html
Joe Armstrong's book "Programming Erlang"

http://learnyousomeerlang.com/ets

### DIFF
--- a/source/languages/en/riak/ops/tuning/erlang.md
+++ b/source/languages/en/riak/ops/tuning/erlang.md
@@ -323,7 +323,7 @@ setting.
 ## Erlang Built-in Storage
 
 Erlang uses a built-in database called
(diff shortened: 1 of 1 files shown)


laurenrother:
:candy:

PR #1664, opened 2015-04-21
Reworked `advanced.config` text

angrycub:
Changed to a more copy/paste friendly version of the text.  This is
still a little problematic in the case where an advanced.config file
already exists, because punctuation has to be handled.


laurenrother:
:cake:

PR #1673, opened 2015-04-27
Updated quickstart to reflect current riak-admin behaviour

RickvdP:
riak-admin is no longer silent when successful, it will give the message below instead

```
Success: staged join request for 'dev2@127.0.0.1' to 'dev1@127.0.0.1'
```

So slightly changed wording to reflect this properly in the docs.

Hope these minor PR's I'm making here and there aren't annoying.


laurenrother:
:candy:

PR #1685, opened 2015-05-12
fix %w array on using/2i page

bkerley:
Ruby %w arrays don't use commas for element delimiting, just whitespace:

```ruby
# commas appear in array elements
%w{ val4, val4, val4a, val4b }
#=> ["val4,", "val4,", "val4a,", "val4b"]

# no commas in array elements
%w{ val4 val4 val4a val4b }
#=> ["val4", "val4", "val4a", "val4b"]
```


laurenrother:
:candy:

PR #1688, opened 2015-05-16
Formatting fix + stats wording

glickbot:
The header for 'Erlang Resource Usage Metrics' was showing up as  'Erlang Resource Usage Metrics These are system metrics from the'

Changed "Tested and Verified Support for Riak 2.x Stats." to "Tested and Verified Support for Riak 2.x.", on Nagios's banner, because Nagios doesn't support stats.

Changed some text in the "Note on Riak 2.x Statistics Support" banner. Reading "full set" would make me think the tools include *all* available stats. However, none of the tools include all the stats, on purpose ( it would be too much info for most tools ). Also, calling out the tools that "support Riak 2.0" would make me think the other tools do not, so I added "we've verified". The text was changed from:
"Look for banners calling out the tools that do support the full set of Riak 2.x statistics!"

to

"Look for banners calling out the tools we've verified that support the latest Riak 2.x statistics!"

This could probably be worded better.


laurenrother:
This looks awesome! Thank you! :candy:

PR #1689, opened 2015-05-18
Update benchmarking.md

xiaclo:
Fixed a minor omission in the command, as-is it will not run with the following error:

```Error in getopt(params) : 
  "/home/username/bench_results/current/" is not a valid option, or does not support an argument
Execution halted```


laurenrother:
:candy:

PR #1692, opened 2015-05-21
Fix code siblings examples

christophermancini:
http://docs.basho.com/riak/latest/dev/references/http/fetch-object/


laurenrother:
:candy:

laurenrother:
Thanks for pointing out that we can't put multiple code blocks of the same type next to each other, @christophermancini

christophermancini:
Glad to help!

PR #1693, opened 2015-05-22
Added information about Non-SSL connections for Riak Control

angrycub:
Added information to clarify that Riak Control is available over the
HTTP port when authentication is off.  Also added some information up
front about how to determine the listener addresses.


laurenrother:
:candy:

PR #1699, opened 2015-05-28
Fix wrong time format description of storage calculation in Riak CS.

ksauzz:
Even though using riak-cs.schema, the time format of storage schedule is not `"HH:MM"`, but `HHMM`. This is a documentation bug in source code of riak_cs and here. It's my fault :(

The description which `riak-cs config describe 'stats.storage.schedule.$time'` shows will be also fixed by https://github.com/basho/riak_cs/pull/1153


laurenrother:
Thank you for the fix, @ksauzz 
:candy:

PR #1721, opened 2015-06-15
Typo in app.config parameter for setting the maximum Erlang VM port

dannylauca:
In the Port Range section of the Erlang VM Tuning chapter, changed the second parameter in the first app.config tab to inet_dist_listen_max.


laurenrother:
:+1:

PR #1723, opened 2015-06-17
Minor fix in Taste of Riak - Erlang

ctcharlton:
In Reading Objects From Riak, this fails:

```
Val2 =:= binary_to_term(riakc_obj:get_value(Fetched2)).
```

because the fetched object is a bit string, which causes a bad argument exception to binary_to_term.


laurenrother:
:candy:

PR #1728, opened 2015-06-19
Update README.md

laurenrother:
Reduce the README to just the information necessary for people to make changes and submit PRs. The inner workings of the docs are in flux, so I removed all the explanations of that. I also noted that PRs must point at a version branch rather than master, as master is being deprecated/removed.


TheOnlyRew:
:+1:

angrycub:
:+1:

PR #1736, opened 2015-07-01
fix invalid multi_backend configuration

rf:
In fact, if you copy paste what was provided in the docs, you'll get

```
10:43:03.471 [error] You've tried to set multi_backend.bitcask.io_mode, but there is no setting with that name.
10:43:03.472 [error]   Did you mean one of these?
10:43:03.518 [error]     multi_backend.$name.bitcask.io_mode
10:43:03.518 [error]     multi_backend.$name.bitcask.data_root
10:43:03.518 [error]     multi_backend.$name.bitcask.expiry
```


source/languages/en/riak/ops/advanced/backends/multi.md
```diff
@@ -99,7 +99,7 @@ Here is an example of the general form for configuring multiple
 backends:
 
 ```riakconf
-multi_backend.$name.$setting_name = setting
+multi_backend.$name.backend_name.$setting_name = setting
```
angrycub: I would say that we tend to think of settings in riak.conf as their whole dotted name representation, `bitcask.io_mode` for example.  So in this case, I think this edit might be overreaching.

source/languages/en/riak/ops/advanced/backends/multi.md
```diff
@@ -107,8 +107,8 @@ to set LevelDB's `bloomfilter` setting to `off` and the Bitcask
 backend's `io_mode` setting to `nif`, you would do that as follows:
 
 ```riakconf
-multi_backend.leveldb.bloomfilter = off
-multi_backend.bitcask.io_mode = nif
+multi_backend.my_backend.leveldb.bloomfilter = off
+multi_backend.my_backend.bitcask.io_mode = nif
```
angrycub: I agree with the spirit of this edit, but we would want to have two different names for the backends, since the name is what refers to the specific backend.
like:

```
multi_backend.my_leveldb_backend.leveldb.bloomfilter = off
multi_backend.my_bitcask_backend.bitcask.io_mode = nif
```

PR #1750, opened 2015-07-22
Fixed brew install command.

DeRauk:
The original command failed for me, replacing 'devel' with 'devrel' fixed it.


laurenrother:
Excellent catch! Thank you! :candy:

PR #1757, opened 2015-08-05
Add git dependency to Installing Erlang

dmitrizagidulin:
Since our kerl instructions depend on git, added it to the pre-reqs.


laurenrother:
:candy:

PR #1760, opened 2015-08-07
Create BDP branch and main page pre-launch

laurenrother:
Add a partial main page for BDP, update the side nav and index, and create git branch for BDP in anticipation of launch.


TheOnlyRew:
Is there a reason you'd like dataplatform-index.md to be empty? I'm guessing the content in bashodataplatform.md should be there instead?

Additionally, the data/versions.yml file will have to be updated, as it's currently listing the 0.8.0 version of the data platform, rather than the 1.0.0 version.

source/languages/en/dataplatform/side_nav.yml
```diff
@@ -7,4 +7,4 @@ dataplatform:
   - title: "Start Here"
     class: "start"
     sub:
-      - "[[Test Page|Test Data Platform Page]]"
+      - "[[Basho Data Platform|Basho Data Platform]]"
```
TheOnlyRew: Will this correctly point to the index page?

data/versions.yml
```diff
@@ -26,7 +26,7 @@ riak:
 riakcscontrol:
   - ['1.0.0', '1.0.1', '1.0.2']
 dataplatform:
-  - ['0.8.0']
+  - ['0.8.0', '1.0.0']
```
TheOnlyRew: I would just remove the 0.8.0 from this. We never actually pushed that version to S3, so we don't want it to show up on the version picker.
Also, we should bump the `dataplatform: '0.8.0'` in the below `currents:` map.

source/languages/en/dataplatform/side_nav.yml
laurenrother: Eff. Probably not.

TheOnlyRew:
Give that side nav a look in the built version, and (assuming it's correct) :cake:

laurenrother:
:raised_hands:

PR #1770, opened 2015-08-19
Modify project meta-data to be 'data-platform'

TheOnlyRew:


TheOnlyRew:
This PR may actually be ignored in favor of 1761. Depends if we want to include these changes as two merges, or just one.

TheOnlyRew:
Turns out this is incorrect. No need to make these changes.

PR #1773, opened 2015-08-21
Update release-notes.md

ph07:
* Added known issue about spurious Spark-Master processes, discovered at final testing
* Added that recommendation against co-locating BDP services on the same box might be dropped in future versions.
* Clarified that 'storage_backend' parameter is located in riak.conf

### DIFF
--- a/source/languages/en/dataplatform/release-notes.md
+++ b/source/languages/en/dataplatform/release-notes.md
@@ -58,7 +58,8 @@ You can find the install packages for BDP [here](LINK) and the instructions to w
 
 ##Known Issues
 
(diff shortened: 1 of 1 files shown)


laurenrother:
Thanks for the additions, Pavel. I've incorporated them here: https://github.com/basho/basho_docs/pull/1774

I incorporated them separately because: 
1.) In general, adding notes about future releases or promises to fix things comes off as overly apologetic and brings up more questions than it answers. If you're using a product, you want to know what it is currently doing, not what is planned for it to do in the future (unless, of course, it's a bug). I've kept some wording about future work in the note about the spark-master processes.
2.) The wording around the spark-master processes was a bit confusing. 

Thanks again!

PR #1787, opened 2015-08-25
Correcting naming of New Relic Plugin

mjbrender:
Typo that @dmitrizagidulin noticed just seconds too late.


mjbrender:
Nvm, you got it on #1786

PR #1790, opened 2015-08-25
Clarifications to Using Data Types page

dmitrizagidulin:
* Remove sentence that's confusing for Java users
* Standardize bucket/bucket type usage for Ruby snippets


laurenrother:
:candy:

PR #1792, opened 2015-08-26
Add Upgrading section

laurenrother:
Add instructions for BDP beta users to upgrade to 1.0.0

### DIFF
--- a/source/languages/en/dataplatform/upgrading.md
+++ b/source/languages/en/dataplatform/upgrading.md
@@ -1,0 +1,44 @@
+---
+title: Upgrading
+project: dataplatform
+version: 1.0.0+
+document: guide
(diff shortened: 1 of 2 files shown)


source/languages/en/dataplatform/upgrading.md
```diff
@@ -0,0 +1,44 @@
...
+version: 1.0.0+
+document: guide
+toc: false
+index: true
+audience: beginner
+---
+
+#Upgrading from Basho Data Platform Beta to 1.0.0
+
+>**NOTE:
```
erjohnson: <div class="note">
<div class="title">Note:</div>
This page only applies to people previously using the beta releases of Basho Data Platform.
</div>

PR #1795, opened 2015-08-27
Update upgrading.md

ph07:
Suggest to make this sub-title smaller, as currently it conflicts with the page title


TheOnlyRew:
-1
We should just change the `title:` of this page to be "Upgrading from Basho Data Platform Beta to 1.0.0", and remove the h1 element.

laurenrother:
Yep. @TheOnlyRew is correct. The header should just be removed entirely and the title should be changed. This will ALSO need to be changed in the sidenav, so I will make this change myself.

PR #1803, opened 2015-08-28
Remove 'bash' from console output snippets

ph07:
Most pages use 'bash' code syntax to show Unix shell commands, but for output just 'code' md is used, This page was not consistent.


TheOnlyRew:
+1
This is very low priority, but good catch.

laurenrother:
Covered by #1810

PR #1809, opened 2015-08-28
Fix URL for Azure install script

roooms:
The current URL returns a 301 redirect as GitHub sends the request on to raw.githubusercontent.com, breaking the install instructions.


laurenrother:
Excellent catch, thank you! :candy:

PR #1810, opened 2015-08-28
Fixed commands following initial launch

paegun:
* fixed commands to be more accurate and consistent throughout.
* added link to vagrant bdp setup as this can help customers who want to give BDP a spin or are running into incorrect or incomplete documentation.


source/languages/en/dataplatform/installing.md
```diff
@@ -61,26 +62,28 @@ If you are using or plan to use Spark, you must install Java 8 on **each node th
...
 If you are on Ubuntu, run the following to install Java 8:
 
-```Ubuntu
-$ sudo apt-get install software-properties-common
-$ sudo apt-get install python-software-properties
-$ sudo add-apt-repository ppa:webupd8team/java
-$ sudo apt-get update
-$ sudo apt-get install -y oracle-java8-installer
+```bash
+sudo apt-get install -y software-properties-common python-software-properties debconf-utils
```
laurenrother: Why remove the $? (As written tone is hard, I want to say this is a completely earnest question out of curiosity.)

source/languages/en/dataplatform/learn-about-dataplatform/leader-election-service.md
```diff
@@ -48,7 +48,7 @@ You can replace “internal” with whatever name you like. You can also specify
 
 For example:
 
-```riak.conf
+```config
```
laurenrother: Also completely earnest question, why 'config' rather than 'riak.conf'?

source/languages/en/dataplatform/using-dataplatform/configuration/setup-a-cluster.md
```diff
@@ -239,37 +242,36 @@ The hostnames you enter must match what you set as `SPARK_MASTER_IP` in priv/spa
...
 
-```
-data-platform-admin add-service-config my-redis redis HOST="10.0.0.1" REDIS_PORT="9999"
+```bash
+sudo data-platform-admin add-service-config my-redis redis HOST="0.0.0.0" REDIS_PORT="»REDIS_PORT«"
 ```
 
 Or you can use the default port by leaving out the `REDIS_PORT` parameter:
 
 ```
```
TheOnlyRew: This should be ```` ```bash````?

source/languages/en/dataplatform/installing.md
paegun: Leading $'s were requested to be removed for consistency, especially with existing Riak KV documentation, see http://docs.basho.com/riak/latest/ops/tuning/open-files-limit/

source/languages/en/dataplatform/learn-about-dataplatform/leader-election-service.md
paegun: The markdown format is ```<lang> .  The syntax highlighting of a config is correct here.  Within the text above, the context of this being riak.conf is set with the reader.

source/languages/en/dataplatform/using-dataplatform/configuration/setup-a-cluster.md
paegun: Yes, this is shell and for consistency sake, we have more ```bash than ```shell throughout our documentation.  I believe we should be using ```shell as our scripts are supposed to be /bin/sh friendly, not using bashisms.  However, consistency is king.  If we want to fix everything to be ```shell, then that should be done across the board, including Riak.  This PR is for BDP only.

source/languages/en/dataplatform/learn-about-dataplatform/leader-election-service.md
TheOnlyRew: So it's known, ```` ```riakconf```` was added to our HTML rendering logic to make sure the syntax highlighting and titles are correct for riak.conf files, and is what should be used to open such code-blocks. See https://github.com/basho/basho_docs/blob/master/lib/sitemap_render_override.rb#L295-L314 .

source/languages/en/dataplatform/using-dataplatform/configuration/setup-a-cluster.md
TheOnlyRew: The only instances of ```` ```shell```` are in BDP actually, and that concern has been raised in another issue. ```` ```bash```` is what's used to denote shell-style code-blocks in the docs everywhere else, and the rendered title is, in fact, "Shell".
So yes, "```` ```bash````" is a bit of a misnomer, but it's only visible in the source.

laurenrother: Wait, @TheOnlyRew , so should we always use ```bash so it shows up as "Shell" (and using ```shell doesn't work)?

laurenrother:
:fish_cake:

source/languages/en/dataplatform/using-dataplatform/configuration/setup-a-cluster.md
TheOnlyRew: That is correct.

Using ```` ```shell```` will sometimes work, but if we want consistent and correct syntax highlighting, we gotta use ```` ```bash````.
Or modify our render code. Which we could totally do. Just not as a part of this PR.

PR #1811, opened 2015-08-30
Correctly render the Data Platform BashoMan

TheOnlyRew:
I'm not sure why linking from /shared/latests/ isn't working, but manually
changing that HTML url property to /shared/1.0.0/ is working. So let's
get that up there.


laurenrother:
:+1:

PR #1848, opened 2015-10-05
WIP: Create top-level installing section for Riak KV

erjohnson:
- Move content from `riak/{VERSION}/ops/building/installing/` to `riak/{VERSION}/installing/`.
- Add 'Installing' section to side navigation.

### DIFF
--- a/data/downloads_gen.yml
+++ b/data/downloads_gen.yml
@@ -1,7 +1,962 @@
 ---
 riak:
+  2.1.1:
+    debian:
+      versions:
+        '6':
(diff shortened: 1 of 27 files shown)


laurenrother:
Let's remove the links to the 5 Minute Install, as it is confusing to me why we would link to another installation process from an installation page. 
Also, I think we should just link to the community projects for installing Riak with Chef/Cloudsoft/etc., since there's no guarantee from us that those will be up to date.
Otherwise, ++ :+1:

PR #1858, opened 2015-10-13
[RFC] Add sections on Riak CS GC cli & configuration

erjohnson:


kuenishi:
This is great, @erjohnson

PR #1866, opened 2015-10-23
Add Riak S2(CS) 2.1.0 and Stanchion 2.1.0 into Version Compatibility Table

ksauzz:
This adds new row about Riak S2 2.1.0 into [Version Compatibility Table](http://docs.basho.com/riakcs/latest/cookbooks/Version-Compatibility/). According to the releasenote, it might be better to include Riak 2.0.5 as well.

##### A quotation from Releasenote

> This is a backwards-compatible* release that introduces a new metrics system, garbage collection refinements, and several other new features. Riak S2 2.1 is designed to work with both Riak KV 2.0.5+ and 2.1.1+.


erjohnson:
Added in via https://github.com/basho/basho_docs/commit/1d66980973153ceb11c0b9d29ea6df5733af0083

ksauzz:
I couldn't find the row of CS 2.1.0 on [the table](http://docs.basho.com/riakcs/2.1.0/cookbooks/Version-Compatibility/) ... Have you deployed your change?

PR #1939, opened 2016-01-29
fix incorrect name of loglevel in Benchmarking

jvf:
The name of the `warn` loglevel was incorrect, it should be `warning`. Setting `warn` in the config leads to keeping the default `debug` level.


laurenrother:
:candy:

PR #1945, opened 2016-02-10
Fixed Solr search schema version

lastzero:
The version must be 1.5 for Riak, not 1.0


lastzero:
This is still not fixed - at least the documentation for 2.1.3 that is currently online, it says

```
<schema name="schedule" version="1.0">
```

See http://docs.basho.com/riak/kv/2.1.3/developing/usage/search-schemas/

PR #1969, opened 2016-03-25
Remove references/links to Riaknostic

laurenrother:


jpease:
To clarify, we don't want to remove all mention of the functionality. We specifically want to remove any links out to riaknostic.basho.com.

So for example, in the commit against source/languages/en/riak/installing/verify-install.md I think most of that content could stay... just kill the link.

laurenrother:
I'm pulling this over to Hugo.

PR #1983, opened 2016-04-07
Fix url link

paultannenbaum:
Having `www` in the url breaks the link. This PR removes and also adds a trailing slash to be consistent with the other urls.

Not sure why there are whitespace differences in this diff, but here is the [actual changeset](https://github.com/basho/basho_docs/compare/riakts/1.2.0...paultannenbaum:link-fix?w=1).


laurenrother:
:candy: Thanks!

PR #1995, opened 2016-04-11
Riak KV 2.1.4

erjohnson:


TheOnlyRew:
👍

laurenrother:
:+1: but I'd like Drew to check it too just  in case.

PR #1996, opened 2016-04-11
Updates to Redis Add-on

laurenrother:
Fixes some broken things post-docs site launch. Addresses issue #181 in private_docs

PR #1997, opened 2016-04-12
Technical Updates to Redis Add-on #1996

paegun:
Addresses RRA technical aspects of issue #181 in private_docs .

* added bucket type note to using-rra.md .
* added brief version of bucket:key commentary from using-rra.md into developing-rra.md.
* added "In the Lab" note regarding ansible example setup to setup-rra.md.


laurenrother:
:+1:

PR #2006, opened 2016-04-13
Fix canonical_link's location into header on Security page

ksauzz:
See: http://docs.basho.com/riak/kv/2.1.4/using/security/


laurenrother:
:+1:

PR #2014, opened 2016-04-14
Update redirects to include malphormed url scheme

erjohnson:


TheOnlyRew:
I'd definitely check to make sure that riak/kv/2.1.4/using/security/best-practices.md, riak/kv/2.1.4/using/security/v2-v3-ssl-ca.md, and riak/kv/2.1.4/using/reference/failure-recovery.md are included in the config.yml IgnoreFiles list, elsewise those pages will be rendered as empty entries.
+1 aside from that, though.

PR #2018, opened 2016-04-19
Remove redundant aliases, fix aliases within configuration section

erjohnson:


TheOnlyRew:
101 files is a very strange number... but I trust your work, so you have my +1.

erjohnson:
The extra is the removal of one of the designating admin user page aliases, I'll fix that up

PR #2034, opened 2016-04-23
Fix incorrect link path #2033

### LINKED ISSUE #2033: incorrect path for planning link [JIRA: DOC-318]
There is a link to the planning page with an incorrect path. The path ends up as /installing/using/planning but it looks like it should be /using/planning/.
This issue is present on http://docs.basho.com/riak/ts/1.2.0/installing/debian-ubuntu/ including all versions and operating systems.

oliryde:
Should fix #2033 (DOC-318)


laurenrother:
Ah, yes! Thank you! 🍬

PR #2040, opened 2016-04-28
totol_leveldb_mem_percent needs to be moved to eleveldb sections

joecaswell:
this setting is ignored in a multi-backend config, it must be in the top level eleveldb section to be effective


laurenrother:
👍 🍥

PR #2041, opened 2016-04-28
fix broken 'creating-objects' link

jordan0day:


laurenrother:
👍 Good catch, and thanks! 🍬

PR #2042, opened 2016-04-28
Clean up redirects, content/ directories

TheOnlyRew:
content/riakcs/ and content/riakts/ were bad directories, and I felt bad for adding them.
This PR removes them, and replaces the redirects those directories were created for with static/riakcs/index/html and static/riakts/index/html .


laurenrother:
🎉 🎊 👐

PR #2050, opened 2016-05-05
Update download generation logic

TheOnlyRew:
And add the TS 1.3.0 source package to the metadata.
And remove packages we'll never host from the metadata.


erjohnson:
👍

PR #2060, opened 2016-05-10
Markdown Fix

paultannenbaum:
- Fixes incorrect markdown formatting for [this page](http://docs.basho.com/riak/ts/1.3.0/using/querying/) that was causing too large of code block to be rendered, with markdown inside that code block (see screenshot).

*** I do not have the whole build process installed on my machine, so probably worth pulling down and doing a build to make sure it is rendering correctly before accepting this PR.

screenshot 2016-05-10 14 33 14


laurenrother:
👍 Thanks!

PR #2062, opened 2016-05-10
add new section covering percent-encoding for url defined keys. rewor…

christophermancini:
…ked endpoint table a little bit for readability.


laurenrother:
So this is happening with the new table:
screen shot 2016-05-11 at 10 18 57 am

While I like the layout, I wonder if having both Method and Type doesn't end up taking up more space and not being informative enough to make up for it.

Alternatively, what happens if we remove the html formatting element and use ` ` by themselves?

laurenrother:
Fixed by https://github.com/basho/basho_docs/pull/2065

PR #2063, opened 2016-05-11
Add KV-related operations to TS Spark-Riak Connector docs

erjohnson:


laurenrother:
I think this content might be confusing in a TS context. Let's discuss.

PR #2064, opened 2016-05-11
Improve .note styles

TheOnlyRew:
I got tired of not fully understanding the difference between blockquotes and class="note" divs, so I did some digging, put together a {{% note %}} shortcode, fussed with css, and captured the attached screenshot that shows what the difference between the different options.

This also corrects the incredibly broken note div that was being displayed on http://docs.basho.com/riak/cs/latest/cookbooks/configuration/reference/

note pane examples


laurenrother:
👍 🍻

PR #2067, opened 2016-05-12
Remove 'fullsync' from description of MDC support

ksauzz:
Riak TS 1.3 doesn't only support fullsync, but also realtime sync. This description looks like realtime sync is not supported. [The both replication have been tested.](https://github.com/basho/riak_test/pull/1034)


laurenrother:
Thanks, @ksauzz ! 👍 🍬

PR #2069, opened 2016-05-12
Install TS from Source

laurenrother:
Add instructions for installing TS from source, and instructions for installing Erlang.


erjohnson:
👍 lgtm

PR #2073, opened 2016-05-12
Use `draft: true` instead of IgnoreFiles

TheOnlyRew:
Important to note that the `draft: true` only reliably works when its the first option in the front matter.

My local testing has shown this change does prevent the draft files from being compiled into the public/ directory, and prevents their menu entries from appearing elsewhere on the site. Confirmation of both of these observations would be appreciated, prior to a merge.

Also, this shaved bout 15 seconds off of my build time, which made me pretty happy.


laurenrother:
👍 I ran a local version (and yeah, faster! WHOO!) and didn't see any pages I knew should be drafts that snuck through. 

@erjohnson When you get a moment, give it a whirl and see if you spot any?

erjohnson:
looks good! 💯

PR #2082, opened 2016-05-20
Fix supercluster settings, and add its cuttlefish style example

ksauzz:
[According to cuttlefish schema](https://github.com/basho/riak_cs/blob/2.1.1/rel/files/riak_cs.schema#L677-L700), supercluster_members settings should be in riak_cs section. Also, I'd like to add cuttlefish style example of riak-cs.conf and stanchion.conf.


erjohnson:
👍

PR #2086, opened 2016-05-23
Fix Downloads anchors

TheOnlyRew:
The root issue was not giving our Markdown parser a chance to generate \<h3> IDs. Without them, our anchorify logic defaulted to "undefined" as the ID name, and everything broke. This is solved by using the Hugo Template bif `markdownify` to generate header tags in the downloads.html _default layout.


TheOnlyRew:
fixes #2084

erjohnson:
👍

PR #2088, opened 2016-05-24
Product Advisory RSS feed

TheOnlyRew:
Some additional details;
- A `copyright:` string was added to config.yaml that's included as part of the rss feed.
- All of the Community pages were given a `date:` entry in their front matter to allow for a valid `<pubDate>` in the RSS feeds. We should consider including a `date:` on all pages, really.
- All product advisory pages were given an entry in the new `"announcements"` taxonomy to all for easy aggregation. (See below,)
  
  ```
  ---
  title: "Socket reuse issue with Riak Golang client 1.5.1"
  description: ""
  date: "2016-03-01"
  announcements: ["Product Advisories"]
  . . .
  ```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -75,8 +75,10 @@ end
 
 ########
 # Build
-desc      "Compile Compressed JS, Compile Compressed CSS, Build Hugo"
+desc      "Compile & Compress JS, Compile & Compress CSS"
 task      :build => ['clean', 'build:js', 'build:css']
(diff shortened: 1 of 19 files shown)


laurenrother:
@TheOnlyRew , this was really great work and thank you very much for doing it. We opted not to RSS feed it based on user research (which, admittedly, we should have done PRIOR to doing all the work 😬 ).

Issue #2090, opened 2016-05-26
Tags visible in Inspecting a Node page on Chrome, explanation also incorrect [JIRA: DOC-390]


nerophon:
https://docs.basho.com/riak/kv/2.1.4/using/cluster-operations/inspecting-node/#riak-admin-status

Paragraph on performance. See attached picture. Also, in fact executing status more than once a minute can have a performance impact even in later versions of Riak, because of the bug that makes stats requests hold back transfers. Customers are advised by CSEs to check stats every 90 – 120 seconds because of this bug.

![screen shot 2016-05-26 at 12 56 34](https://cloud.githubusercontent.com/assets/4341829/15573809/69a75d36-2341-11e6-9c47-77a5b5a441ae.png)

Basho-JIRA:
https://github.com/basho/basho_docs/pull/2091 

***[posted via JIRA by Eric Johnson]***

PR #2091, opened 2016-05-27
Fix malformed link, remove version tags, add note on stats performance

erjohnson:
#2077 (DOC-376) #2090 (DOC-390)


laurenrother:
👍 🍐

PR #2092, opened 2016-05-27
Remove quanta requirement from Planning and Table Arch

laurenrother:
Per a May 26th conversation in HipChat and email, a quantum is not required in the partition key. Removed requirement language from the relevant docs.


erjohnson:
👍 💯

PR #2094, opened 2016-05-31
fix ports and URLs

kesslerm:
The rest of the documentation uses Riak's default HTTP port number 8098, and expresses HTTP URLs in the form `types/<name>/buckets/<name>` or `buckets/<name>` instead of the antiquated `riak/<name>`


laurenrother:
This is excellent, @kesslerm ! Thank you!

PR #2096, opened 2016-06-02
Fix generated formatting of version strings

kesslerm:
Hugo generates a fraction (7/2) from 2.0.7/2.1.4.


laurenrother:
@erjohnson and @TheOnlyRew Pinging you on this just so you both know ^^^

PR #2097, opened 2016-06-02
Use riak.conf for MDC anti-entropy configuration

kesslerm:
We configure anti-entropy via riak.conf in other places,
and there is a danger to use inconsistent configuration
if the two methods are mixed.


laurenrother:
👍 🐧

PR #2098, opened 2016-06-03
Fix rendering of 24 / 7

kesslerm:
The default Hugo configuration renders 24/7 as a fraction.
Add additional spaces to make this look more natural in the text.


laurenrother:
👍 🍃

PR #2113, opened 2016-06-16
Fix installation instructions for stanchion on Ubuntu/Debian

kesslerm:
Don't state to install Riak CS, when actually installing
the stanchion package.


laurenrother:
👍 🍐

PR #2116, opened 2016-06-20
Added note about write-once bucket bypassing hooks

angrycub:
Since the write-once path bypasses the put_fsm and doesn't fire the
hooks on its own, they are completely ignored.  The main concern being
that realtime replication uses postcommit hooks to enqueue objects;
therefore, realtime can not work with write-once buckets.  Fullsync
works as expected.


laurenrother:
👍 🍬 

Thanks @angrycub !

PR #2117, opened 2016-06-20
Changed template to allow markdown in title attribute

angrycub:
This change will cause the contents of the title attribute to be markdownified.  The main upside is to allow for backticked items appearing in note titles to be drawn correctly.

### DIFF
--- a/layouts/shortcodes/note.html
+++ b/layouts/shortcodes/note.html
@@ -1,6 +1,6 @@
 <div class="note">
   {{ if .Get "title" }}
-    <div class="title">{{ with .Get "title" }}{{.}}{{ end }}</div>
+    <div class="title">{{ with .Get "title" }}{{ markdownify . }}{{ end }}</div>
   {{ end }}
   {{ .Inner }}
 </div>


laurenrother:
@TheOnlyRew Can you take a look at this and +1, please (and thank you)?

TheOnlyRew:
Will do. I intend to push this change a bit, and see how bad it looks when misused, so it may be a couple days before I'm done.

laurenrother:
Just gonna bump this on your radar @TheOnlyRew .

TheOnlyRew:
First I would like to apologize for dragging my feet on this. Time is hard.

Below are some notes I took  while messing around with this change. Overall I'm really happy with the change. Enough that I think we should make a similar change markdownify page titles as well as note titles.

> shortcodes don't allow newlines in parameters, even when wrapped in multi-line strings (\`backticks\`), so most of crazy crap is actually impossible.
> 
> Our current style sheets mess up \_emphasized\_ test in titles, but that should get fixed as part of my ongoing work.
> 
> Horizontal bars (`-----`) actually look pretty good... though you can't add any text.
> 
> Prepending four spaces in front of text ("&nbsp;&nbsp;&nbsp;&nbsp;like this") will actually make the title a code block.
> 
> Images actually look pretty okay in titles... who would have thought?
> 
> HTML tags can be added in titles, so we will be able to shoot ourselves in the foot. But that's really not new.

PR #2118, opened 2016-06-20
Corrected 2.1.1 to match other versions

angrycub:
During this correction, it was observed that the notes do not allow reference style links from the main document, presumably due to scope issues. I had already applied the changes to one version.  This fixes 2.1.1 to look like the other versions.  -cv


TheOnlyRew:
:+1:

Sorry that took so long

PR #2120, opened 2016-06-22
Remove munchkin.marketo kruft from markdown content

TheOnlyRew:


laurenrother:
👍 
Note that this was the script we were asked to add 4/28/15. I think one year is long enough to maintain a marketing campaign and then drop it unless otherwise informed.

PR #2125, opened 2016-06-27
Fix typos in "Implementing a Document Store"

ches:
I can check older versions to see if these fixes are also needed there, just quickly filed this initially using the GitHub web editing where it's convenient to change one file.


erjohnson:
👍

PR #2142, opened 2016-06-29
MapUpdate parameter missing in UpdateMap.Builder()

Rilussion:
Fix might need to propagate to prior versions too


laurenrother:
👍 🍬 
Thanks @Rilussion !

PR #2143, opened 2016-06-29
Corrected `stanchion_host` to `listener`

angrycub:
The stanchion.conf file configures the IP address of the stanchion listener using the `listener` value not `stanchion_host`.  This is very likely copypasta.


erjohnson:
👍

PR #2153, opened 2016-07-05
Find/Replace clobbered versions

laurenrother:
#whoops 
Went back and updated 1.3.1 to 1.3.0 where appropriate. Also added in the actual date of release. #genius


TheOnlyRew:
👍 Good catch, and sorry I completely missed that in the initial review.

PR #2161, opened 2016-07-11
[secondary-indexes]: fix mapreduce key count example

kesslerm:
In the current form, the (bash) shell will substitute
'$bucket' with the empty string (in most cases, unless
a '$bucket' variable already exists). The operation then
fails with

An error occurred parsing the inputs field.
[{unknown_field_type,<<"some_bucket_name">>}]

where the field type is the name of the first bucket
ever created.


laurenrother:
👍 🍬

PR #2164, opened 2016-07-12
Correctly link from the configuration reference to backend configuration references

TheOnlyRew:
**NOTE** that we may want to consider re-adding backend configuration references to /configuring/reference.md rather than linking to the specific backend page as this change does. I think it would be nice to have a single, canonical page that documents all major configuration options, though it is useful to discuss in detail the ramifications of any given configuration option.

Any other opinions?


laurenrother:
👍 💃

PR #2167, opened 2016-07-14
Fix broken links in 2.0.7 & 2.1.4, add concepts/strong-consistency, add dynamo paper

laurenrother:
Broken links from previous versions of the docs have been passed along as each release comes out. This PR updates broken links in the two most recent versions of Riak KV to avoid passing them along. Broken links were identified through Google Webmaster tools Crawl Error report and Integrity Plus broken link checker.

This PR also adds back Strong Consistency under Concepts (under Learning) because a.) it should really be there and b.) there are a LOT of links to it.

This PR also adds back the Dynamo paper and reformats it. The Dynamo paper was removed during the Hugo transition because it was unreadable on some screens and the side-by-side layout was unreactive. It has been reformatted to be simpler and the commentary to be more straightforward.


erjohnson:
👍 💯

Issue #1840, opened 2015-09-28
ulimit still listed as 65,000 for 2.x


binarytemple:
http://docs.basho.com/riak/2.0.6/ops/tuning/open-files-limit/

Current recommendation is to increase the ulimit to 200,000

Basho-JIRA:
Is this still the case? 

***[posted via JIRA by Eric Johnson]***

PR #2168, opened 2016-07-15
Change ulimit increase on Open Files Limit doc

erjohnson:
per https://github.com/basho/basho_docs/issues/1840

Changes only applied to KV 2.1.4. Will apply changes to the remaining KV pages (2.0+) once this is 👍


laurenrother:
👍

PR #2169, opened 2016-07-18
Update Riak CS installation docs

roooms:
- Remove reference to installing Erlang for Riak KV as this only applies to source and is referenced correctly during Riak KV source install
- Reference correct Riak KV version for each Riak CS version
- Remove version compatibility table and instead link to version-compatibility page
- Change ‘Riak’ to ‘Riak KV’ where appropriate
- Remove reference to specific install directories as this can vary a lot between installations


laurenrother:
👍 🎉 🍊 This is so excellent! Thanks, @roooms !

PR #2179, opened 2016-07-22
Set the Media Type on more file extensions

TheOnlyRew:


TheOnlyRew:
After spending what I'm going to delicately call 'entirely too much' time trying to rewrite content types, I'm pretty sure the fastest way to make this change is going to be to delete everything, and rewrite all of it. Again.

This is something that will have to happen pretty soon anyway, taking into account the meta noindex tags being added to the archived content, and the change in version selectors that's going to touch every current page.

This change is safe to merge now, by the by. New and updated objects will have media types set, and all other content will remain unaffected.

laurenrother:
🍊 👻 👍 🎱 🔮

PR #2180, opened 2016-07-26
Upgrade from Ruby 2.2.3 to 2.2.5

TheOnlyRew:
There was a vulnerability warning issued for Ruby 2.2.3, so we're
upgrading versions, and taking the chance to upgrade all gems used.

To perform the upgrade (assuming rvm is in use (which it should be ))

    rvm install ruby-2.2.5
    # verify the correct ruby is in use, `which ruby` should do
    gem install bundler
    bundle install


laurenrother:
🎉

PR #2183, opened 2016-07-30
[JS Version-Selector - PART 1] Version Selector root element and SCSS style

TheOnlyRew:
Create a version_selector.html partial next to -- rather than in place of -- version_ddown.html, and give it some fresh SCSS in version-selector.scss.

The plan is that this root-level element will be expanded upon using runtime javascript (PR pending). The end result of that hypothetical JS should look something like the HTML currently commented out in version_selector.html, and will have the same functionality as the current version drop down.

The naming conventions used here are what I'm targeting for the larger-scale HTML and SCSS rewrite that's still incoming. Until that change set lands, the naming will be inconsistent, but I consider that a small price to pay for not duplicating work (I should be able to drop this change set into the larger change as-is).

> **NOTE**: The structure of this (and related) PR(s) is a bit of an experiment
> I'm running.
>
> I'm kinda dead sick of making a huge number of tangentially related changes,
> then requesting a review on an unreadable mass of diffs spanning most of a
> codebase. This time, I'm breaking this larger change set into logically
> grouped subsets -- each with its own focused PR -- and "stacking" those PRs.
> The hope is that because each PR contains tightly coupled changes (and is
> substantially smaller than whole), reviewers will be better able to give more
> focused and through feedback.
>
> None of these related PRs should be merge until they've all been given the
> green light. Once that's happened, and all changes that need to be made are
> re-reviewed, we can collapse the whole stack from the top all the way to
> master.


laurenrother:
Insofar as I can tell exactly what's going on here, it looks good to me. 👍

erjohnson:
👍

PR #2191, opened 2016-08-04
[JS Version-Selector - PART 5] Add `version_history` front matter where applicable

TheOnlyRew:
The most significant portions of this change are the update to
layouts/partials/common_variables.html and layouts/partials/head.html to
incorporate the version_history map into `<head><meta>` tags that will expose
the other changes (adding `.in` to some TS pages that were added after 1.0.0) to
the JS.

> **NOTE**: The structure of this (and related) PR(s) is a bit of an experiment
> I'm running.
>
> I'm kinda dead sick of making a huge number of tangentially related changes,
> then requesting a review on an unreadable mass of diffs spanning most of a
> codebase. This time, I'm breaking this larger change set into logically
> grouped subsets -- each with its own focused PR -- and "stacking" those PRs.
> The hope is that because each PR contains tightly coupled changes (and is
> substantially smaller than whole), reviewers will be better able to give more
> focused and through feedback.
>
> None of these related PRs should be merge until they've all been given the
> green light. Once that's happened, and all changes that need to be made are
> re-reviewed, we can collapse the whole stack from the top all the way to
> master.


layouts/partials/head.html
```diff
@@ -2,9 +2,23 @@
   <meta charset="utf-8">
   <meta content="IE=edge,chrome=1" http-equiv="X-UA-Compatible">
   <meta name="viewport" content="width=device-width">
+
   <meta name="project" content="{{ .Params.project }}">
   <meta name="version" content="{{ .Params.project_version }}">
-  <meta name="base-url" content="/{{ .Scratch.Get "site_relative_url" }}">
+
+  <!-- TODO: Adding these leading slashes is... pretty dumb. IMO. We should
```
laurenrother: Leading slashes in the sense of having to put /riak vs riak/ >>>project<<</version?

laurenrother:
Big 👍 . That was a lot of work! Thanks, Drew!

layouts/partials/head.html
TheOnlyRew: So in this case, `site_relative_url` is going to be something like index/,

TheOnlyRew: So in this case, `site_relative_url` is going to be something like `riak/kv/2.1.4/installing/` even thought that's not actually a site-relative url. `/riak/kv/2.1.4/installing/` (note the leading `/`) is a site-relative url, and we want to make sure we record the correct thing in the meta tag.

PR #2202, opened 2016-08-15
Misc Build System Improvements

TheOnlyRew:
The big change here is the to s3_deploy script, and the way it verifies the correctness of our archived content. Should be safer, or at least more explicit in its log messages.
We're also cleaning up the Guardfile some, which has been neglected for a little while and may deserve a bit more love.


laurenrother:
Looks good, Drew!

erjohnson:
👍 lgtm!

PR #2206, opened 2016-08-18
Corrected Read quorum tables

angrycub:
Per JIRA DOC-299, updated the table to reflect the default values of `R`
and `notfound_okay`


laurenrother:
👍 🍬

PR #2211, opened 2016-08-23
Fix broken links in KV latest(s)

laurenrother:
Except the MapReduce erl links because we're working out what those should be.


erjohnson:
👍

PR #2222, opened 2016-08-26
Fix wait-for-service code example, add note to check riak.conf on TS 1.4.0 upgrading

erjohnson:


laurenrother:
👍 Thanks so much for the quick fix, @erjohnson !

PR #2226, opened 2016-08-30
Fix the issue where h# elements covered preceeding text

TheOnlyRew:


TheOnlyRew:
Forgot to note in the comment that @angrycub came up with the code to correct the behavior. I just put it in the .scss file.

angrycub:
Tested this PR locally with hugo and this... 
works on my machine

laurenrother:
👍 @TheOnlyRew did you want @erjohnson to look this over before we merge?

erjohnson:
👍

PR #2229, opened 2016-08-31
Replace Riak KV logo with higher resolution copy

roooms:
Sourced from http://basho.com/design-assets-hub/ this is the png cropped (by eye) to line up with the much larger but resized Riak TS logo. This doesn’t satisfy my OCD for having it perfect, but it looks 100x better than the low-res version used at the moment.

Tested locally with `hugo server`.


laurenrother:
👍 Works on my laptop. :)

PR #2234, opened 2016-09-09
Great Rework Part 04 -- CSS classes for controlling `display:`, `float:`, etc.

TheOnlyRew:
> The classes defined below directly affect the flow of the HTML documents.
> Basic styles and class-specific styles should not alter the `display`,
> `float`, or `clear` properties. If an element needs to have those properties
> changed, one of the below classes should be applied alongside any other
> custom styles.

This is a style of building flow that I'm not 100% sure of, but feel strongly about trying out. After all, if it's a common trick to include `<div style="clear: both">`, wouldn't it make both the CSS and the HTML more clear to precede those clearfixes with `<div style="float: left">`? (My answer is "yes", in case there was confusion.) Currently this _formatting.scss doesn't include positional information, which... I'm torn on. I would like to encode `position: fixed` in the HTML, same as floats and clears, but I'm suspicious that writing reactive CSS may not mesh well with that encoding scheme. I intend to revisit that specific question later, once the rework is in a more complete state.

---

**THIS PR IS A WORK IN PROGRESS**
I've opened this as part of a set of stacked PRs so that I can more cleanly and more quickly share the work that I'm doing. Any changes made to this branch will likely come in the form of a force push, so local updates should be performed with

```
git checkout great_rework/04/formatting_scss
git fetch origin great_rework/04/formatting_scss
git reset origin/great_rework/04/formatting_scss --hard
```

Please feel free to comment on anything you see here; offer suggestions, provide corrections, raise concerns. the whole point of this exercise is to engender communication.

But _**do not merge this PR**_.

### DIFF
--- a/dynamic/css/base/_formatting.scss
+++ b/dynamic/css/base/_formatting.scss
@@ -1,0 +1,141 @@
+/**
+ * Formatting-Specific classes
+ * ===========================
+ * The classes defined below directly affect the flow of the HTML documents.
+ * Basic styles and class-specific styles should not alter the `display`,
+ * `float`, or `clear` properties. If an element needs to have those properties
+ * changed, one of the below classes should be applied alongside any other
+ * custom styles.
+ *
+ * We're not including classes designed to hide or partially hide elements here
+ * because Bootstrap provides those for us,
+ *     .invisible       -- `display: hidden`
+ *     .hidden-print    -- `display: none` in print media
+ *     .hidden-*-up     -- `display: none` when screen is equal/larger than *
+ *     .hidden-*-down   -- `display: none` when screen is equal/larger than *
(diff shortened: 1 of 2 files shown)


TheOnlyRew:
After looking at how much I munged the history of this change set, and after looking at how tightly coupled and _big_ this change set is, I'm throwing in the towel with this experiment.

I think it may have worked if there was iterative review to go along with the iterative progress but I failed to secure that kind of review, and I let the stacked PRs fall out of date with the head of development. At this point, there's too much entropy to make cleaning the history a worthwhile task.

Moving forward I'm going to consider building a "false history" that will (hopefully) accomplish the goals of this stacked PR set; smaller, logically grouped change sets that make reviews possible.

PR #2245, opened 2016-09-13
Corrected alter-user commands that had typos

angrycub:
Some of the instances of calls to riak-admin were missing `security` in the command. This PR fixes those.


erjohnson:
👍

PR #2252, opened 2016-09-23
Dbrown riak ts fix

laurenrother:
Dan noticed some things amiss with the TS example table and some of the client examples, so this PR covers:
- Removing erroneous comma causing errors when creating the GeoCheckin table.
- Update example response from status request to match the updated table creation command used for Riak TS 1.4.
- Add missing column from example table: The region column is not in this example snippet but is in most others. The inconsistency would cause issues for anyone following along.

### DIFF
--- a/content/riak/ts/1.4.0/using/planning.md
+++ b/content/riak/ts/1.4.0/using/planning.md
@@ -40,6 +40,7 @@ In order to create a working Riak TS table, you'll need to plan your table out. 
 CREATE TABLE GeoCheckin
 (
    id           SINT64    NOT NULL,
+   region       VARCHAR   NOT NULL,
    time         TIMESTAMP NOT NULL,
    region       VARCHAR   NOT NULL,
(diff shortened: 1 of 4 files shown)


content/riak/ts/1.4.0/using/planning.md
```diff
@@ -40,6 +40,7 @@ In order to create a working Riak TS table, you'll need to plan your table out.
 CREATE TABLE GeoCheckin
 (
    id           SINT64    NOT NULL,
+   region       VARCHAR   NOT NULL,
```
alexmoore: This is a duplicate column (see column after `time`), was this intentional?

alexmoore: Actually, line 45 is the duplicate, this one should stay.

laurenrother:
Thanks @pogzie @roooms & @alexmoore !

PR #2256, opened 2016-10-04
Add appropriate flag for downgrading to rpm command

paulhenrich:
I just saw that this was necessary while doing downgrade testing on Riak KV. There are no CentOS instructions for KV, but I found the TS instructions, which needed an additional flag.


erjohnson:
👍

PR #2258, opened 2016-10-06
Add commercial only flag to all of the SNMP pages

angrycub:


content/riak/kv/2.0.2/using/reference/snmp.md
```diff
@@ -9,7 +9,10 @@ menu:
     identifier: "managing_ref_snmp"
     weight: 107
     parent: "managing_ref"
-toc: true
+toc: truecommercial_offering: true
```
erjohnson:
```
toc: true
commercial_offering: true
```

PR #2259, opened 2016-10-06
Updated the example table and sample codes

pogzie:
Standardized the examples to the following values (id, region, state, time, weather, temperature)

```
1, 'South Atlantic', 'Florida', 1451606401, 'hot', 23.5
2, 'East North Central', 'Illinois', 1451606402, 'windy', 19.8
```

Note on the time: The time is the Unix time equivalent of January 1, 2016 00:00:01 and 00:00:02 respectively. This is a better example than the ambiguous 1234567 and standardizes the Python example that uses an actual date and time function. 

Examples tested and updated: Java, Ruby, Python, Javascript, Erlang
Examples not updated: C#, PHP, Go


erjohnson:
💯 👍

Issue #2272, opened 2016-10-27
force-replace command typo [JIRA: DOC-695]


glickbot:
http://docs.basho.com/riak/kv/2.1.4/using/admin/commands/#force-replace

says:
riak-admin force-replace <node_being_replaced> <replacement_node>

should be:
riak-admin cluster force-replace <node_being_replaced> <replacement_node>

Basho-JIRA:
https://github.com/basho/basho_docs/pull/2273 

***[posted via JIRA by Eric Johnson]***

PR #2273, opened 2016-10-27
Fix DOC-695 - force-replace command typo

angrycub:
`riak-admin force-replace` should be `riak-admin cluster
force-replace`

Fixes Issue #2272 (DOC-695)


erjohnson:
👍

PR #2278, opened 2016-11-04
Add configuration and metadata to v3 operations

shaneutt:
This adds documentation on the preferred way to handle propagation of search indices and bucket properties in Replication.


erjohnson:
👍

PR #2299, opened 2016-11-17
Great Rework Part 17 -- Small but important fixes and improvements

TheOnlyRew:
This is a bit of a catch-all PR that includes a pretty large number of relatively small changes to style, format, layout, comments, etc.. The highlights are;

* Reorganized the _colors.scss file to make room for by descriptive (`$gray-40`) and semantic (`$font-color`) color variables.
* Moved away from `$gray-dark` and `$gray-lighter`, etc. in favor of going purely descriptive. This was done with the intent to come back later and move away from purely descriptive colors in code.
* Correctly opened the <html> element, so modernizr.js can actually work now.
* Started using modernizr's `.js` vs `.no-js` class in styles.
* Pretty drastically re-styled the look of anchors in primary.

---
**THIS PR IS A WORK IN PROGRESS**
I've opened this as part of a set of stacked PRs so that I can more cleanly and more quickly share the work that I'm doing. Any changes made to this branch will likely come in the form of a force push, so local updates should be performed with

```
git checkout great_rework/17/small_fixes_improvements
git fetch origin great_rework/17/small_fixes_improvements
git reset origin/great_rework/17/small_fixes_improvements --hard
```


TheOnlyRew:
After looking at how much I munged the history of this change set, and after looking at how tightly coupled and _big_ this change set is, I'm throwing in the towel with this experiment.

I think it may have worked if there was iterative review to go along with the iterative progress but I failed to secure that kind of review, and I let the stacked PRs fall out of date with the head of development. At this point, there's too much entropy to make cleaning the history a worthwhile task.

Moving forward I'm going to consider building a "false history" that will (hopefully) accomplish the goals of this stacked PR set; smaller, logically grouped change sets that make reviews possible.

PR #2309, opened 2016-11-21
Correct Process

jenagraham:
Correct the upgrade/downgrade documentation process.  Particularly, replace the command that brings the Yokozuna service back up on a node after the AAE trees have been rebuilt.  The current command binds the pid to the `riak attach` shell rather than the Yokozuna process which causes the process to be unintentionally down'ed again when the shell is exited.


erjohnson:
👍

PR #2312, opened 2016-11-24
Clarify documentation on usage of data types via HTTP API

lucafavatella:


content/riak/kv/2.2.0/developing/api/http.md
```diff
@@ -47,7 +47,8 @@ Method | URL | Doc
 ## Riak-Data-Type-related Operations
 
 For documentation on the HTTP API for [Riak Data Types](/riak/kv/2.2.0/learn/concepts/crdts),
-see the `curl` examples in [Using Data Types](/riak/kv/2.2.0/developing/data-types).
```
lucafavatella: A simpler patch would be linking to http://docs.basho.com/riak/kv/2.2.0/developing/data-types/#usage-examples like:
```
see the `curl` examples in [Using Data Types](/riak/kv/2.2.0/developing/data-types/#usage-examples).
```
but I am not sure whether that would work because I found no instance of `relativeurl/#foo` in the docs, hence I went for the straightforward option.

laurenrother:
🍰 👍 Thanks!

PR #2317, opened 2016-11-28
Ease consultation of HTTP data types API for advanced users

lucafavatella:
See current technical documentation in [edoc](https://github.com/basho/riak_kv/blob/a9ad5f2547d1f185eced3dd72f2eb65c70108f6e/src/riak_kv_wm_crdt.erl#L22-L120).

It looks like Basho publishes built edoc for client components e.g. [riakc](https://basho.github.io/riak-erlang-client/) but not for server components like riak_kv (that makes sense).


laurenrother:
🍭 👍

PR #2329, opened 2016-12-08
Updated the Replication Hook API page

angrycub:
Fixed a typo in the module name in the sample, moved to reference style links, and moved the API up to the top


laurenrother:
https://www.myptsd.com/c/attachments/scared_cat_0-jpg.16785/

PR #2331, opened 2016-12-12
Add CentOS to RHEL package title.

laurenrother:
Per DOC-654, add CentOS to RHEL heading on downloads page to make the package compatibility clear. And change SUSE to SuSE.


TheOnlyRew:
lgtm.
Consider changing "SUSE" to "SuSE" as per @gordonguthrie's correction?

PR #2341, opened 2016-12-19
Added note about AAE requirement for YZ

angrycub:
Riak search uses AAE to help maintain consistency between Riak and Solr.

### DIFF
--- a/content/riak/kv/2.0.0/developing/usage/search.md
+++ b/content/riak/kv/2.0.0/developing/usage/search.md
@@ -29,6 +29,13 @@ to properly store and later query for values.
 3. **Bucket-index association** signals to Riak *when* to index values
    (this also includes bucket type-index association)
 
+{{% note %}}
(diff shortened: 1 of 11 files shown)


TheOnlyRew:
I'm a little late here, but I'm not sure the language "Riak Search requires" is quite correct. It's certainly better to have AAE enabled, but YZ should be able to interact w/ Solr just fine w/o AAE.

Mostly just wanted to comment on the "Turtley Awesome"ness that's going on here....

angrycub:
In talking it over with Engineering, the risk of divergence (especially with the configurable batching queue blowoffs) is too great to call it not required without a user very carefully understanding the tradeoffs.  We could weaken the wording to "strongly encourage"; however, that doesn't quite convey the level of encouragement that we would actually give the users to use AAE with YZ.  It would only be in fairly special cases (with a lot of caveats) that we would have a user run it otherwise.

PR #2342, opened 2016-12-19
Technically AAE is not "required" for Riak search

angrycub:
Per the discussion in https://github.com/basho/basho_docs/pull/2341, changing "is required" to "should not be disabled" will be more accurate.  Added a few more words for color, ideally to dissuade people from disabling AAE in search use cases willy-nilly.  This is to address the concerns raised by @TheOnlyRew in https://github.com/basho/basho_docs/pull/2341


TheOnlyRew:
Fully addresses my concerns. +1 from me.

laurenrother:
👍 🎉

PR #2358, opened 2016-12-21
Fixed links in ts/1.5.0/releasenotes.md

glickbot:
The following links were relative ( riak/ instead of /riak/ ):
http://docs.basho.com/riak/ts/1.5.0/releasenotes/riak/ts/1.5.0/using/querying/select/order-by
http://docs.basho.com/riak/ts/1.5.0/releasenotes/riak/ts/1.5.0/using/querying/select/limit
http://docs.basho.com/riak/ts/1.5.0/releasenotes/riak/ts/1.5.0/using/querying/delete

Also, Atom "helpfully" removed some whitespace at the end of some lines? Not sure if that's going to mess anything up.


erjohnson:
👍

PR #2360, opened 2016-12-22
Add warning about latency with INSERT

laurenrother:
Per DOC-698, add warning to Writing Data doc about latency with INSERT.


erjohnson:
👍 lgtm

PR #2364, opened 2016-12-30
Fix a couple of links in conflict resolution page

lucafavatella:
This is a minor patch for fixing a couple of links in https://docs.basho.com/riak/kv/2.2.0/developing/usage/conflict-resolution/#timestamp-based-resolution

I did not test that the HTML rendering now works as expected, but I think I did not make it any worse. (I checked that the referred entities exist, though.)


laurenrother:
Thanks, @lucafavatella :)

PR #2370, opened 2017-01-05
Update riak control auth.mode setting

erjohnson:
Should this be applied to any other version prior to 2.1.4?


angrycub:
This should apply to versions 2.0 and greater.

angrycub:
:+1:

PR #2373, opened 2017-01-09
WIP - link checker rake task

erjohnson:
Looking for some eyes on this and any feedback. Still a few open TODOs on this, notably:

* Have output go to a file (which would be ignored by git)
* Add a progress bar using that progressbar gem while the task is running

Eventually the plan is to use the Crawler class to plug in further tasks (such as a spell checker).

@TheOnlyRew


erjohnson:
Changes to the where the links are pulled from (ie ignoring the side nav) led to a need for more robust exception handling / validation of URLs as URL-like strings were being pulled from code examples.

I've addressed this but will do further testing / tinkering to make sure it doesn't break and re-submit the PR.

PR #2374, opened 2017-01-09
Update to use ProgressBar 1.8.2

TheOnlyRew:
Four years seems a little long to wait for an upgrade.
NOTE: This has not been tested in a full deploy as of yet.


TheOnlyRew:
Note, this will require a `bundle install` to upgrade your local gem set.

erjohnson:
👍

PR #2380, opened 2017-01-13
Great Rework -- Aggregate

TheOnlyRew:
This is a maybe-dummy PR to show the entire set of changes in one place.

Might be a bit hard to review.


christophermancini:
So, I can't speak for all the changes to the CSS, JS or other assets, however the PR builds both the assets and site via hugo. It looks good to me.

PR #2393, opened 2017-01-24
Query info update

laurenrother:
1. Update /configuring/riakconf - Per DOC-756, ran through defaults and information in Configuring riak.conf page and updated where needed and clarified where I could.
2. Query guidelines and configuration update - Per DOC-756, update the guidelines page, table architecture page, and link to the configuration page from LIMIT.


erjohnson:
👍

PR #2404, opened 2017-01-24
Great Rework -- Review Test Part 4 -- Blanket JS/Coffee update

TheOnlyRew:
This is a test to see how possible it is to parse up reviews for a change set that touches 50k lines (#2380). These different "parts" are going to be stacked, so the changes are cumulative. Mostly, they're so we can logically separate different change sets to (theoretically) simplify the review process. To run the code and build the actual website, the great_rework/aggregate branch should be used.

---

Like it says on the tin. This might be too big to review. I guess we'll see.

### DIFF
--- a/dynamic/js/README.md
+++ b/dynamic/js/README.md
@@ -1,0 +1,58 @@
+JavaScript and Friends
+======================
+
+This directory should contain all of the code necessary to run JS code on the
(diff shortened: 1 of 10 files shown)


TheOnlyRew:
This PR was effectively merged in as part of the branch `great_rework/aggregate`.

I don't think we're going to get any reviews here, so I'm going to go ahead and close out this  all associated PRs.

PR #2409, opened 2017-01-25
correct total query size formula + new section on inmem max

hmmr:
Here's a couple of items related to https://github.com/basho/riak_kv/pull/1609 and https://github.com/basho/riak_kv/pull/1587.

1. Correct the descriptive variable name in the expression used to compute the total query size. This is to reflect the fix in code that will go into 1.5.2 (and also to 1.6 as a matter of course).
2. Add a new section describing the newly introduced parameter `riak_kv.query.timeseries.qbuf_inmem_max`.


laurenrother:
👍 🍰

PR #2423, opened 2017-02-03
Reorg and pretty-up dynamo.md

TheOnlyRew:
Like it says on the tin.

This may need to be re-applied to any pending versions. 2.0.8?


laurenrother:
👍 🍰 :tumbler_glass:  

It does. I can backport it Monday.

PR #2425, opened 2017-02-06
Integrate Store Object in HTTP API overview

lucafavatella:


laurenrother:
Thanks @lucafavatella !

PR #2426, opened 2017-02-06
Hide menubars on the index page

TheOnlyRew:
This commit also un-hides main.js and main.css from .gitignore, so keeping static files up to date should be easier.


erjohnson:
👍

PR #2428, opened 2017-02-06
Safari flicker fix

TheOnlyRew:
Like it says on the tin.

Note that this commit is stacked on top of #2427, but only because they both modify static content, and I don't want to deal with the conflict that would cause.


erjohnson:
💯 👍

PR #2436, opened 2017-02-10
Add clarification to `riak-admin cluster plan`

laurenrother:
Per Magnus, `cluster plan` can perform a wide range of actions that are not immediately clear. Add this info to the docs.


laurenrother:
Sorry, Eric, I couldn't get this to resolve down and I figured it's all going to the same place so whatevs.

erjohnson:
👍

PR #2437, opened 2017-02-10
Add note about manual URL requirement

laurenrother:
With the site redesign, the way the redirects from the splash page work has changed and is limited to/dependent on the S3 bucket. Which means that, in order to get from the splash page to the KV/TS/CS index page, you must manually enter version number in the URL. 

Alternatively, you can sit around thinking it/your laptop is broken like I did. But I don't recommend this approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Please let us know if you'd like to help out!
 
 >**Heads-up**
 >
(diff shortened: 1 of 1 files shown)


TheOnlyRew:
:+1:

PR #2441, opened 2017-02-13
Separate tiered instructions from non-tiered

laurenrother:
Broke out the Healing Corrupted LevelDBs section into steps for tiered storage and steps for regular LevelDB. Edited for word choice, brevity, and clarity.


angrycub:
:+1:  looking good to me.

PR #2442, opened 2017-02-13
Disable compilation of Bootstrap's JS library

TheOnlyRew:
We're not explicitly using Bootstrap's JS library anywhere and I did a quick scan of the Bootstrap CSS and saw no indication that any of it relied on JS. I rebuilt and ran the updated JS locally, and saw no issues with the site. That's not to say I did an exhaustive search, but all major functionality that I could find was in place and unaffected by the removal.


erjohnson:
👍

PR #2453, opened 2017-02-16
Remove field from map preferring documented list syntax

lucafavatella:
Please refer to commit message for details.

Tested on Riak KV 2.2.0.


lucafavatella:
It is preferable to put list syntax in example so that user reading page knows how to remove multiple fields at once.

christophermancini:
Thanks @lucafavatella.  @laurenrother this looks like a reasonable change.

PR #2457, opened 2017-02-18
Fix url for riak 2.2 source.

lo5an:


laurenrother:
Nice catch. :) Thank you!

PR #2458, opened 2017-02-18
describe inverse distrib functions (PERCENTILE etc)

hmmr:
RTS-1749, RTS-1746, RTS-1740.

Describe new aggregation functions `PERCENTILE_CONT`, `PERCENTILE_DISC`, `MEDIAN` and `MODE`, as implemented in https://github.com/basho/riak_kv/pull/1624. Slated for 1.6.0.

### DIFF
--- a/content/riak/ts/1.6.0/using/querying/select/aggregate-functions.md
+++ b/content/riak/ts/1.6.0/using/querying/select/aggregate-functions.md
@@ -23,6 +23,7 @@ aliases:
 
 
 [arithmetic]: ../arithmetic-operations
+[riak.conf]: /riak/ts/1.6.0/configuring/riakconf/
 
 
 You can turn a set of rows in your Riak TS table into a value with the aggregate feature. This document will walk you through the functions that make up aggregation in Riak TS.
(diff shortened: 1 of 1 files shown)


content/riak/ts/1.6.0/using/querying/select/aggregate-functions.md
```diff
@@ -143,3 +146,29 @@ Returns `NULL` if no values were returned or all values were `NULL`.
...
+
+### MEDIAN
+
+Equivalent to `PERCENTILE_DISC(x, 0.5)`.
+
+### MODE
+
+Calculate the mode (i.e., the value occurring with the highest frequency) of observations in the selection, with nulls discarded.  If there are more than one modes, the lowest one is returned.  The column argument must be of numeric type (`sint64`, `double` or `timestamp`).  Return type is the type of the column argument.
+
+{{% note title="Notes on inverse distribution functions" %}}
```
gordonguthrie: Notes don't work in this dialect of Markdown...

laurenrother: They do work in Hugo though :)

hmmr: I couldn't be wrong as they did exist elsewhere in the document..

content/riak/ts/1.6.0/using/querying/select/aggregate-functions.md
```diff
@@ -143,3 +146,29 @@ Returns `NULL` if no values were returned or all values were `NULL`.
 |-------------------|-------------|
 | sint64            | sint64 |
 | double            | double |
+
+### `PERCENTILE_DISC`, `PERCENTILE_CONT`
+
+Calculate a percentile, given as a value in the range [0..1], in a population of entries in the selection, with null values discarded.  The column argument must be of numeric type (`sint64`, `double` or `timestamp`).  Return type is the type of the column argument for `PERCENTILE_DISC`, and `double` for `PERCENTILE_CONT`.
```
macintux: Unless the value can be the integer 0 or 1, I would suggest expressing the range as `[0.0..1.0]`.

macintux: wait, now I'm confused. If it's an integer, it must be `[0..100]`?

hmmr: The DISC version always returns the value of the actual cell. It does so without casting it to some other type, and by that token, without double promotion.

The CONT version, on the other hand, does an interpolation step, which _does_ necessitate a double promotion. Which is why it is always 'double'.

(Or could it be that you are confused about the `x` and `0.3` in `PERCENTILE(x, 0.3)`? `x` is the argument column, `0.3` is the function positional parameter.)

macintux: I got confused about the purpose of the range. I was thinking that would be the output, not the input.

hmmr: Using floats to indicate percentile parameter range in 3ed8000.

PR #2459, opened 2017-02-20
Fix typo

lucafavatella:
The typo fix is correct I believe.

Off-topic: I am a bit surprised that the PBC client needs to request the body for getting the key - I would expect that the client gets the key in response if not specified in request regardless from body.


laurenrother:
👍 🍰

PR #2468, opened 2017-03-02
Install instructions for KV 2.0.8 for FreeBSD 10

TheOnlyRew:
Also, remove a note about KV 1.2.


erjohnson:
Looks good to me 👍

laurenrother:
Should this be applied to 2.2.0?

TheOnlyRew:
something similar should perhaps be applied, but I've not looked into the specifics of the version numbers.

PR #2471, opened 2017-03-03
Fixes download urls

angrycub:
As part of the version docs migration, we didn't catch the x.y part of the download URLs


erjohnson:
👍

PR #2472, opened 2017-03-03
Update python.md

jenagraham:
Update conflict resolution code examples and comment explaining the example to include an object store after the reload call.


erjohnson:
👍

PR #2475, opened 2017-03-09
Moved Sources for Generated Public/Static Assets

paegun:
* fixed hugo upgrade to 0.19 regression
  * moved the RRA uml deployment diagram sources to public_src.
  * added a note to the README about the process of generating from such sources
    * including the nudge to make the generation repeatable via a script.


laurenrother:
Thanks so much for this @paegun !

PR #2478, opened 2017-03-10
Add per table object expiration draft

erjohnson:
On the fence as to whether this should be a separate page, but as a separate page we can add in alter table instructions (for enabling on existing tables or disabling per table), avoid any redirect soup with regard to renaming the global-object-expiration page to reflect the inclusion of per table instructions, etc.


laurenrother:
Closing in favor of PR #2490

PR #2483, opened 2017-03-13
Mikedotson/mdc bucket properties

mikedotson:
Update Documentation for MDC Bucket Properties override


laurenrother:
This is super incredible, @mikedotson ! Thank you so much!

PR #2488, opened 2017-03-14
Smarg/misc errata

BillSmargiassi:
Contains 3 cherry-pickable *minor* errata:

1. An "i.e." that should be "e.g.". Since Latin abbreviations are confusing (case in point), I chose "for example".
2. TS docs using "CO2" as an example, slipped in an "O2".
3. A pair of swapped links to command line documentation for `riak-admin` and `riak`.


laurenrother:
These are awesome @BillSmargiassi ! Thanks!

PR #2491, opened 2017-03-14
WIP add group by time example

erjohnson:
Not sure if the given example is correct. Also thinking we may want to expand this into it's own page similar to `IN(...)` as it can also be used with `SELECT`


laurenrother:
The time function can be used with `SELECT`?

PR #2493, opened 2017-03-15
Review bucket types in HTTP API object pages

lucafavatella:
I was using DELETE and I found I had to prepend `/types/<type>` in order to make it work (I had written the key as datatype). The HTTP API overview page lists only the URLs with the `/types/<type>` prefix - **not sure if the specific pages shall list both** (I think yes).
I **did not proofread the modified pages for inconsistencies with bucket types**.


laurenrother:
Thanks, @lucafavatella !

PR #2507, opened 2017-04-14
Warn Riak KV user of potential newlines in log messages

lucafavatella:


laurenrother:
Ugh, wow. Good find, @lucafavatella . Thanks for the note!

PR #2509, opened 2017-04-27
Fix erroneous note & spelling error

laurenrother:
When fixing Issue #2090, we misread the error and so left the incorrect information present in the docs. This commit removes the note about a bug (long gone since KV 1.3) and fixes a spelling error.


angrycub:
:+1:

PR #2525, opened 2019-12-15
Remove Homebrew install instuctions

maxnordlund:
It has been removed [1][] due to Erlang 17 dependency, which only supports EOL OpenSSL. The formulae also had numerous issues building, which most likely needs upstream changes (aka here) before being submitted again. An alternative would be an official tap [2][].

This PR only removes the now inaccurate docs, as the other changes are probably much bigger and definitely requires some thinking.

[1]: https://github.com/Homebrew/homebrew-core/pull/46552
[2]: https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap


maxnordlund:
The build failure is unrelated to this PR, (ruby version mismatch between installed and Gemfile).

PR #2527, opened 2021-04-21
Add method for listing buckets via `aae_fold` to V2.9.1 + docs

JMercerGit:
A safe method of listing buckets regardless of backend chosen. Listing buckets had previously not been production safe, but can still be required in production environments - it can now be managed safely via an aae_fold. - Should be added to the docs

See https://github.com/basho/riak/blob/riak-2.9.8/RELEASE-NOTES.md#riak-kv-291-release-notes for more details


JMercerGit:
Added to the wrong branch.

PR #2529, opened 2021-10-07
Fix issue with malformed Markdown code fence breaking formatting

mplewis:
Hello! I was having trouble reading the 2.2.3 Riak KV docs for Go:

https://docs.riak.com/riak/kv/latest/developing/getting-started/golang/object-modeling/index.html

Search on this page for:

```
We'll be using the bucket `Users` to store our data.
```

This 4-tick code fence breaks rendering for a significant part of the page and makes it harder to read, even though it renders fine in GitHub.

This fix should allow this document to render properly on docs.riak.com.

Thank you for the detailed docs and I hope this makes them better for the next readers.


martinsumner (approved):

martinsumner:
I don't think the riak docs are auto-generated from here any more.  This will be a good test to find out

PR #2533, opened 2023-08-24
Massive update to docs

pjaclark:
Creating as a Draft PR so the below issues can be considered before merging.

Contains docs for all released versions of Riak KV since 2.0.0. New docs for Riak KV 2.2.6, 2.9.x, 3.0.x, and 3.2.x, Riak CS 2.1.2, 3.0.x, Riak TS 3.0.x. Includes a multitude of fixes to content, plus layout, render and markdown issues. A live version of this is at [https://www.tiot.jp/riak-docs/](https://www.tiot.jp/riak-docs/).

Issues that should be looked at before merging:
- Search
  - currently uses Google Custom Search tied to the TI Tokyo domain
  - should probably be entirely replaced with an offline search from pre-build JSON indexes
- Values in `config.yaml` (currently TI Tokyo based for want of a better option):
  - GitHub links
  - Contact Us Email
  - Contact Us URL
  - Community Email
  - Docs Contact Email
  - Security Contact Email
  - Example domain (used in code examples)
- Downloads
  - `download_root` is tied to https://files.tiot.jp/riak/ as that is the only source of all compiled packages.
  - The rake download generator uses an SFTP login on sftp.tiot.jp to make the list of downloads
  - The structure is based off the old Basho S3 instance, which no longer works
  - We're fine with continuing to host this as we make all the packages anyway
- Docker scripts
  - We should make one for docs.riak.com (super easy)

The search issue is the only big item that needs looking at for deploying onto another server. This is why I haven't deployed a copy to https://docs.riak.info/ yet.

I suggest we make a set of config.yaml files (e.g. `config.riak-com.yaml` and `config.tiot-jp.yaml`, for example) with specific changes for each. The README would then tell the user to copy one and (if wanted) make changes for their use.


pjaclark:
@nsaadouni Here's the PR you mentioned in [this commit](https://github.com/basho/riak/commit/6ae237d686f2218b9f388d8711dd897a73206447).

pjaclark:
Amongst the other content fixes, I have now added an offline search function using [Lunr JS](https://lunrjs.com/). Using the same search box as before one can search the currently selected version of documentation. This means that if performing a search while browsing the KV 3.2.0 docs, it won't find irrelevant KV 2.0.1 content. This avoids the issue of search results finding `riak-admin` and `riak-repl` for KV 2.x when you need `riak admin` and `riak repl` for KV 3.x and vice-versa.

The offline index is built by Hugo as part of the site build process. The new search function is done purely in the browser, so there is not depency on a 3rd party website to provide a search function. This makes the search function independent of the domain that Riak Docs is on.

For those wishing to build the docs themselves, we have included a docker script that builds an image of Hugo that can then either serve the site directly via Hugo or be used to export all the necessary html files for the site to be hosted via Apache or nginx. Please see _docker/deply.riak-info.docs.sh_ and _docker/docker-compose.generate-docs-riak-info.yaml_ for an example deploy script.

